PR #19, opened 2019-10-07
Longest substring without repeating char. O(n^2) solution. time limit…

manishjangid:
Time limit exceeded. trying to improve it by storing the previous results. but its not helping. Need to think more about it.

### DIFF
--- a/Problem2.cpp
+++ b/Problem2.cpp
@@ -1,0 +1,92 @@
+/* Problem Statement:
+
+3. Longest Substring Without Repeating Characters
+Medium
+
+Given a string, find the length of the longest substring without repeating characters.
+
+Example 1:
+
+Input: "abcabcbb"
+Output: 3 
+Explanation: The answer is "abc", with the length of 3. 
+
+Example 2:
+
+Input: "bbbbb"
+Output: 1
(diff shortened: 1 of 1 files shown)


Problem2.cpp
```diff
@@ -0,0 +1,92 @@
...
+    /*
+    * Approach in detail"
+    * Start with each character in the string and then try to see what is the longest subsequence in the string which has no
+      duplicates. Use hash table to store this info and once we get a duplicate entry in hash. record the length and update it
+      if it is larger than existing one.
+
+      The solution fails for a string of size 31000 due to time limit constraints. 
+     *
+     *  TODO: NEED TO IMPROVE THE ALGORITHM 
+    */
```
kaurramanpreet: you could have solved this question using set by reducing the time complexity
ths time complexity for your solution is O(n^2) and not O(n)

kaurramanpreet: i saw your other solution with O(n) time complexity suing sets
great work!!

manishjangid: @kaurramanpreet  yes, you are right. let me write the correct time complexity in the program.

PR #63, opened 2020-03-23
Strings-1 done in Brute Force.pls do review

Mohan110594:


rajatkumar01:
Time complexity will be O(n+m) as you are running two separate loops . try to optimize the solution for customSortString.

Please revisit the time and space complexity as  in lengthOfLongestSubstring you are running a single loop. So the time complexity will be O(n).

PR #75, opened 2020-03-24
Completed problem 2

PratikshaPP:


Lohith-Aswathappa:
good job Pratiksha!
Try the custom sort problem as well

PR #77, opened 2020-03-24
add custom sort strings and longest non repeating substring - Please review

priyanka1012-singhal:


Problem2.java
```diff
@@ -0,0 +1,30 @@
+// Problem2 : Longest Substring Without Repeating Characters(https://leetcode.com/problems/longest-substring-without-repeating-characters/)
+    //Time complexity - O(n) Space - O(m) where m is the number of unique characters
```
super30admin: Space will be O(1) as there can be max 26 characters in the set if lowercase letters and both upper case and lowercase then 52..Still a constant

super30admin:
Good Job....Please review the inline comment

PR #80, opened 2020-03-24
String 1

ptndoss:


super30admin:
Good Job Doss. Both solutions are correct

PR #85, opened 2020-03-26
Custom Sort and Longest Substring with non repeeating chars

singhperryo1:


LongestSubString.java
```diff
@@ -0,0 +1,108 @@
+/*
+Passed All Test Cases 
+
+HashSet Approach 
+
+Big O 
+Time-> O(N), N is # of chars in string s
+Space-> O(N), N is # of chars in string s
```
super30admin: Space here is O(1) as well. We are just using a hashset. In worst case it will be taking 256 entries. Worst case scenario of time will never be dependent on  N that is length of string

super30admin:
Good job Gagandeep. Please review inline comments

PR #90, opened 2020-05-05
prob-2 cmpleted

schakrad:


Problem-2 cmpleted.java
```diff
@@ -0,0 +1,28 @@
+
+// TC: O(N^2)
```
devashishnyati, line 2: Try to optimize it.

devashishnyati:
Good job Sindhu!
Try to optimize longest substring problem.

PR #102, opened 2020-06-08
Completed Strings-1 - Please review

khushal1198:


Strings_1_Problem_1_customSort.java
```diff
@@ -0,0 +1,51 @@
+import java.util.HashMap;
+import java.util.Map;
+
+//Time Complexity : O(n) ( n = length of t + length of s )
+//Space Complexity : O(n) (n = length of t string)
```
sowmyadvn, line 5: This has to be different from n, as n is assigned to be len(t)+len(s) already. But, complexity is correct

sowmyadvn:
Good job Khushal!

sowmyadvn:
Good job Khushal!

PR #121, opened 2020-06-18
Completed

tejassjain:


longest_substring_without_repeating_characters.java
```diff
@@ -0,0 +1,20 @@
+//Time Complexity:O(n)
+//Space Complexity: O(min(m,n))
```
AroopN, line 2: Please state what is m and n ?

AroopN:
Good job Tejas. Please state what variables used in TC and SC stand for.

PR #149, opened 2020-07-21
Completed Strings-1

Prashanth073-cloud:


AroopN:
Good work, Prashanth. Please also consider space taken by string builder object in space complexity for custom sort.
Also for LongestSubstr, hashmap can only contain only upto 26 alphabets. So, the space is constant.

AroopN:
Good work, Prashanth. Please also consider space taken by string builder object in space complexity for custom sort.
Also for LongestSubstr, hashmap can only contain only upto 26 alphabets. So, the space is constant.

PR #647, opened 2022-08-17
String solutions

pavanreddy21:
- String solutions


LongestSubString.java
```diff
@@ -0,0 +1,20 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n)
```
daredevilteja, line 2: SC is O(1)

daredevilteja:
Good Work! Keep it up

PR #723, opened 2023-01-12
Completed String -1

ManishPandey1508:


CustomSort.java
```diff
@@ -0,0 +1,33 @@
+import java.util.HashMap;
+import java.util.Map;
+//Time Complexity : O(n) 
+//Space Complexity : O(n) for hash map and String Builder
```
osandeepreddy, line 4: Space complexity O(1) for both solutions.

osandeepreddy:
Space complexity O(1) for both solutions.

CustomSort.java
ManishPandey1508: Got it, thanks for correcting it.

PR #822, opened 2023-09-19
Strings-1-Complete

deeksha9107:


Exercise_1.py
```diff
@@ -0,0 +1,29 @@
+# ## Problem1
+# Custom Sort String (https://leetcode.com/problems/custom-sort-string/)
+# // Time Complexity :O(1)
```
ishika2741: Check TC

Exercise_2.py
```diff
@@ -0,0 +1,31 @@
+# Problem2
+# Longest Substring Without Repeating Characters(https://leetcode.com/problems/longest-substring-without-repeating-characters/)
+# // Time Complexity :O(2n)
```
ishika2741, line 3: Submit map solution as well where Tc would be perfect O(n)

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #891, opened 2024-02-15
Completed

Siri-isha:


LongestSubstring.java
```diff
@@ -0,0 +1,31 @@
+import java.util.HashSet;
+import java.util.Set;
+//TC - O(n)
+// SC - O(N) - N is length of resultant longest substring
```
ishika2741: Resultant space does not count in complexity

ishika2741 (commented):
Good work!

ishika2741:
Good work!

LongestSubstring.java
Siri-isha: Yes, modified the SC to O(1)

PR #943, opened 2024-08-06
completed strings-1 v1

dhikki-study:
##791. Custom Sort String################################################
// Time Complexity : m+n
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : no problem

// Your code here along with comments explaining your approach in three sentences only
We created hash map for s string and stored the string and its count, now we go through the order and check if the element is present in hash map if it is present than we will add it to list as the count and once it is added to list we will remove the entry from the hash map. Once this process is complete we will be left with the elements which are not in order and we can add them also in list and at last return the list by performing the join

###3. Longest Substring Without Repeating Characters##################################################
// Time Complexity : 2N
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : no problem

// Your code here along with comments explaining your approach in three sentences only
We used sliding window technique along with set, where we used fast and slow pointer and fast pointer keep going through the list and add elements to set, If any element is again encountered by fast pointer we will move slow pointer till it passes the element which fast has encountered and also along the path it will delete the entry for set also

### DIFF
--- a/3. Longest Substring Without Repeating Characters.py
+++ b/3. Longest Substring Without Repeating Characters.py
@@ -1,0 +1,28 @@
+###253. Meeting Rooms II################################################################################################
+// Time Complexity : 2N
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : no problem
+
+// Your code here along with comments explaining your approach in three sentences only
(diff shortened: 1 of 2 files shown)


Nandha2Tech (commented):
Great Work!

Nandha2Tech:
Great Work!

PR #966, opened 2024-09-16
done

akash-shindhe86:


Erlybird:
Good job ! If possible, when you have multiple solutions for a single problem, create a new file for each solution and commit them instead of having all the solutions in one file.

Erlybird:
Good job ! If possible, when you have multiple solutions for a single problem, create a new file for each solution and commit them instead of having all the solutions in one file.

PR #1021, opened 2025-02-20
Solved Strings-1

rajivranjansahu:


custom-sort-string.java
```diff
@@ -0,0 +1,35 @@
+// Time Complexity : O(M + N)
+// Space Complexity : O(M + N)
```
ishika2741, line 2: Space complexity will be constant as map will consist of constant number of characters

longest-substring-without-repeating-characters.java
```diff
@@ -0,0 +1,22 @@
+// Time Complexity : O(N)
+// Space Complexity : O(N)
```
ishika2741, line 2: same as above

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1083, opened 2025-06-12
Completed String-1

adityajidhonde:


super30admin:
1. Correctness:
   - Problem-1.cpp: The solution correctly implements the custom sort string problem. It uses a frequency map to count characters in 's' and then constructs the result string based on the 'order' string followed by remaining characters. This approach is correct.
   - Problem2.cpp: The solution attempts to solve the longest substring without repeating characters problem using a sliding window approach. However, there is a logical error in the implementation. The right pointer (`r`) and left pointer (`l`) are incremented together, which doesn't correctly handle the sliding window. The logic should maintain a window where all characters are unique.

2. Time Complexity:
   - Problem-1.cpp: The time complexity is correctly stated as O(N), where N is the size of string 's'. This is accurate since we traverse both strings once.
   - Problem2.cpp: The time complexity is incorrectly stated as O(N). The current implementation may not achieve O(N) due to the incorrect sliding window logic. A correct sliding window approach should indeed be O(N).

3. Space Complexity:
   - Problem-1.cpp: The space complexity is correctly stated as O(N) in the worst case (when all characters are unique) and O(1) on average (assuming a limited character set). This is accurate.
   - Problem2.cpp: The space complexity is correctly stated as O(N) in the worst case (when all characters are unique) and O(1) on average (assuming a limited character set). This is accurate.

4. Code Quality:
   - Problem-1.cpp: The code is well-structured and readable. The use of a map to store frequencies is appropriate. The comments explain the approach clearly.
   - Problem2.cpp: The code is less readable due to the incorrect logic. The variable names (`f`, `l`, `r`) could be more descriptive (e.g., `left`, `right`). The comments do not fully explain the flawed logic.

5. Efficiency:
   - Problem-1.cpp: The solution is efficient and follows a standard approach. No major optimizations are needed.
   - Problem2.cpp: The solution needs optimization to correctly implement the sliding window approach. The current logic does not correctly handle cases where characters repeat within the window.

PR #1090, opened 2025-07-01
Strings-1 : Problem 1 and 2 Implemented

amardkumar:


super30admin:
1. Correctness:
- For CustomSortString.java: The solution correctly implements the custom sorting by first counting characters in 's' and then appending them in the order specified by 'order'. The remaining characters are appended at the end. This appears correct.
- For LongestSubstring.java: The solution correctly implements the sliding window approach to find the longest substring without repeating characters. The logic for updating the left pointer is correct.

2. Time Complexity:
- Both solutions correctly state their time complexities as O(n), which is accurate for both problems.

3. Space Complexity:
- CustomSortString.java correctly states O(n) space complexity due to the HashMap usage.
- LongestSubstring.java states O(1) space complexity, which is technically correct assuming a fixed character set (ASCII), but could be clarified as O(min(m,n)) where m is the character set size.

4. Code Quality:
- Both solutions are well-structured and readable.
- Variable names are meaningful (map, sb, left, right, max, etc.).
- Could benefit from more comments explaining the algorithm steps.
- In CustomSortString.java, the initial if condition in the first loop could be simplified using map.getOrDefault().

5. Efficiency:
- Both solutions are efficient implementations of standard approaches.
- For CustomSortString.java, the nested loops for appending characters could be replaced with String.repeat() in Java 11+ for slightly cleaner code.
- For LongestSubstring.java, the solution is already optimal.

Potential Edge Cases:
- CustomSortString.java should handle empty strings for both inputs.
- LongestSubstring.java should handle strings with all identical characters.

super30admin:
The student's solution demonstrates good understanding of the problems and provides correct implementations for both problems. Here's a detailed evaluation:

1. CustomSortString.java:
   - Correctness: The solution correctly implements the custom sort by first counting characters from 's' and then appending them in the order specified by 'order'. Any remaining characters are appended at the end.
   - Time Complexity: Correctly identified as O(n) where n is the length of string 's'.
   - Space Complexity: Correctly identified as O(n) for the HashMap storage.
   - Code Quality: The code is well-structured and readable. One minor improvement could be to use map.getOrDefault() when initializing counts.
   - Efficiency: The solution is efficient. No major optimizations needed.

2. LongestSubstring.java:
   - Correctness: Correctly implements the sliding window approach to find the longest substring without repeating characters.
   - Time Complexity: Correctly identified as O(n) where n is the length of the string.
   - Space Complexity: Correctly identified as O(1) since the character set is limited (though technically O(min(m,n)) where m is character set size).
   - Code Quality: The code is clean and efficient. Variable names are appropriate.
   - Efficiency: The solution is optimal. The sliding window approach is the most efficient for this problem.

General suggestions:
1. Consider adding brief comments explaining the algorithm's approach.
2. For CustomSortString, the initial count initialization could be simplified using getOrDefault().
3. Edge cases are handled well in both solutions.

PR #1093, opened 2025-07-01
Done Strings 1

ivnat07:


super30admin:
1. Correctness:
   - For `customSortString`: The solution correctly implements the approach of using a HashMap to count character frequencies and then building the result string based on the order specified. It handles remaining characters appropriately.
   - For `longestStringWithoutRepeatingChars`: The sliding window approach with HashSet is correctly implemented to find the longest substring without repeating characters.

2. Time Complexity:
   - `customSortString`: Correctly stated as O(n) where n is the length of string s.
   - `longestStringWithoutRepeatingChars`: Correctly stated as O(2n) (which is O(n)) due to the sliding window approach.

3. Space Complexity:
   - Both solutions correctly state O(1) space complexity (assuming a fixed character set), though technically it's O(k) where k is the number of unique characters.

4. Code Quality:
   - Good use of StringBuilder for efficient string concatenation.
   - Proper variable naming and clear logic flow.
   - Could improve with more comments explaining the algorithm steps.
   - The second solution could use better variable names (e.g., 'maxans' could be 'maxLength').

5. Efficiency:
   - Both solutions are efficient for their respective problems.
   - For `longestStringWithoutRepeatingChars`, the solution could be optimized to O(n) with a single pass by storing character indices in a HashMap and directly moving the left pointer when duplicates are found.

6. Edge Cases:
   - Both solutions handle edge cases well (empty strings, all unique characters, etc.).
   - Could explicitly mention these edge case considerations in comments.

super30admin:
1. Correctness:
   - For `customSortString.java`, the solution correctly uses a HashMap to count character frequencies and then constructs the result string based on the order specified. This approach is correct and handles the problem requirements well.
   - For `longestStringWithoutRepeatingChars.java`, the solution uses a sliding window approach with a HashSet to track characters, which is a standard approach for this problem. It correctly handles the case of finding the longest substring without repeating characters.

2. Time Complexity:
   - For `customSortString.java`, the time complexity is O(n + m), where n is the length of `s` and m is the length of `order`. The student's claim of O(n) is slightly off because it doesn't account for the iteration over `order`.
   - For `longestStringWithoutRepeatingChars.java`, the time complexity is O(2n) as stated, which simplifies to O(n). This is accurate for the sliding window approach.

3. Space Complexity:
   - For both solutions, the space complexity is O(1) if we consider the character set to be fixed (e.g., ASCII characters). This is correct.

4. Code Quality:
   - The code is generally well-structured and readable. However, there are minor improvements that could be made:
     - In `customSortString.java`, the variable name `st` could be more descriptive, such as `result`.
     - In `longestStringWithoutRepeatingChars.java`, the variable name `maxans` could be more descriptive, such as `maxLength`.
     - The comments are helpful but could be more detailed to explain the logic more clearly.

5. Efficiency:
   - Both solutions are efficient and use standard approaches. No major optimizations are needed, but the student could consider using an array instead of a HashMap for `customSortString.java` if the character set is known to be small (e.g., lowercase letters only), which could slightly improve performance.

PR #1097, opened 2025-07-01
done Strings1

puneetGupts:


super30admin:
1. Correctness:
   - For `customSortString`, the solution correctly implements a counting approach to sort characters in `s` based on the order specified in `order`. It handles characters not in `order` by appending them at the end, which is correct.
   - For `lengthOfLongestSubstring`, the solution uses a sliding window approach with a hash map to track the last seen index of each character. This is a standard and correct approach for this problem.

2. Time Complexity:
   - `customSortString`: O(n + m), where n is the length of `s` and m is the length of `order`. This is optimal.
   - `lengthOfLongestSubstring`: O(n), where n is the length of `s`. This is optimal.

3. Space Complexity:
   - `customSortString`: O(n) for the hash map and the result list. This is optimal.
   - `lengthOfLongestSubstring`: O(min(m, n)), where m is the size of the character set. This is optimal.

4. Code Quality:
   - The code is generally well-structured and readable. Variable names are descriptive (`hmap`, `res`, `charset`, etc.).
   - The commented-out solutions in `LongestSubstringWithoutRepeatingChar.py` could be removed to avoid clutter, as they are not being used.
   - The `customSortString` method could benefit from a docstring explaining its purpose and parameters.

5. Efficiency:
   - Both solutions are efficient and use standard approaches. No major optimizations are needed.
   - For `lengthOfLongestSubstring`, the condition `if s[r] in charset` could be combined with the subsequent line for brevity, but this is a minor stylistic choice.

PR #1099, opened 2025-07-01
Strings-1

Mounika8300:


super30admin:
1. Correctness:
   - For `CustomSort.java`, the solution correctly implements a frequency map and builds the result string based on the given order, then appends remaining characters. This approach is correct for the problem.
   - For `LargestCommonSubString.java`, the solution correctly implements the sliding window technique to find the longest substring without repeating characters. The logic for updating the left pointer and calculating the max length is correct.

2. Time Complexity:
   - `CustomSort.java`: The time complexity is correctly stated as O(n), where n is the length of string s.
   - `LargestCommonSubString.java`: The time complexity is correctly stated as O(n), where n is the length of string s.

3. Space Complexity:
   - `CustomSort.java`: The space complexity is O(1) because the hashmap size is bounded by the number of unique characters (26 for lowercase English letters).
   - `LargestCommonSubString.java`: The space complexity is O(1) for the same reason as above.

4. Code Quality:
   - Both solutions are well-structured and readable.
   - Comments are clear and explain the approach well.
   - Variable names are meaningful (e.g., `occurances`, `left`, `right`, `max`).
   - The use of `StringBuilder` in `CustomSort.java` is efficient for string concatenation.

5. Efficiency:
   - Both solutions are efficient and follow standard approaches for their respective problems.
   - No major optimizations are needed, but minor improvements could include:
     - In `CustomSort.java`, the initial check `if(!map.containsKey(c))` is unnecessary because `map.put(c, map.getOrDefault(c, 0) + 1)` can be used instead.
     - In `LargestCommonSubString.java`, the logic can be slightly simplified by using `Math.max` directly in the return statement without an intermediate `max` variable.

6. Edge Cases:
   - Both solutions handle edge cases well (e.g., empty strings, strings with all unique characters, strings with all repeating characters).

PR #1112, opened 2025-07-02
Completed String 1

chanukyagattu:


super30admin:
1. Correctness:
   - For `CustomSortString_LC791.java`, the solution correctly implements the custom sort by first counting frequencies of characters in `s` and then appending them in the order specified by `order`. The remaining characters not in `order` are appended at the end, which matches the problem requirements.
   - For `LongestSubstringWoRepeatChars_LC3.java`, the solution correctly uses a sliding window approach with a HashSet to track characters and maintain the longest substring without repeating characters. This is a standard approach for this problem.

2. Time Complexity:
   - `CustomSortString_LC791.java`: The student correctly identifies the time complexity as O(N + K), where N is the length of `s` and K is the length of `order`. This is accurate.
   - `LongestSubstringWoRepeatChars_LC3.java`: The student correctly identifies the time complexity as O(n), where n is the length of `s`. This is accurate because each character is processed at most twice (once when added to the set and once when removed).

3. Space Complexity:
   - `CustomSortString_LC791.java`: The student correctly identifies the space complexity as O(N) due to the HashMap storing frequencies. This is accurate.
   - `LongestSubstringWoRepeatChars_LC3.java`: The student correctly identifies the space complexity as O(min(n, m)), where m is the size of the character set. This is accurate because the HashSet size is bounded by the number of unique characters in the string.

4. Code Quality:
   - Both solutions are well-structured and readable. The use of meaningful variable names (`freq`, `sb`, `st`, `len`, etc.) and comments enhances readability.
   - The comments explaining time and space complexity are helpful and follow best practices.
   - The code follows consistent indentation and formatting.

5. Efficiency:
   - Both solutions are efficient and use optimal approaches for their respective problems.
   - For `CustomSortString_LC791.java`, the use of `StringBuilder` is efficient for building the result string.
   - For `LongestSubstringWoRepeatChars_LC3.java`, the sliding window approach is optimal, and the use of a HashSet ensures O(1) average-time complexity for lookups and deletions.

Areas for Improvement:
   - In `CustomSortString_LC791.java`, the inner `while` loop could be replaced with a `for` loop to append characters `count` times, which might be slightly more readable.
   - In `LongestSubstringWoRepeatChars_LC3.java`, the initial check for `s.length() == 0` could be combined with the loop condition to reduce redundancy, though this is a minor point.

PR #1119, opened 2025-07-05
Strings 1 complete

joycesilvera:


super30admin:
Let me evaluate each solution separately.

1. CustomSortString.java:
- Correctness: The solution correctly implements the custom sort by first counting characters from 's', then appending them in 'order' sequence, and finally appending remaining characters. This approach is correct.
- Time Complexity: The stated O(n + m) is accurate where n is length of s and m is length of order.
- Space Complexity: The analysis is correct - O(n) due to the StringBuilder and HashMap.
- Code Quality: The code is well-structured and readable. Good use of HashMap and StringBuilder. The comments are helpful.
- Efficiency: The solution is already efficient. One minor improvement could be to use an array[26] instead of HashMap since we're dealing with lowercase letters (assuming input is lowercase), which would make it O(1) space.

2. LongestSubstringWithoutRepeat.java:
- Correctness: The sliding window approach is correctly implemented to find the longest substring without repeating characters.
- Time Complexity: The stated O(n) is correct as each character is processed at most twice (by left and right pointers).
- Space Complexity: The analysis is correct - O(min(n, m)) where m is character set size.
- Code Quality: The code is clean and efficient. Good use of HashSet and sliding window technique. The variable names are clear.
- Efficiency: The solution is optimal. No significant improvements needed.

General strengths:
- Both solutions demonstrate good understanding of the problems
- Proper use of data structures (HashMap, HashSet, StringBuilder)
- Correct time and space complexity analysis
- Clean, readable code with appropriate comments

Areas for improvement:
- Could add input validation (null checks, empty strings)
- For CustomSortString, could mention assumption about lowercase letters if that's the case
- Could add more test case considerations in comments

PR #1120, opened 2025-07-05
completed

mauliciousware:


super30admin:
Let's evaluate each problem separately.

Problem 1: Custom Sort String
1. Correctness: The solution correctly implements the custom sort by first counting characters in 's' and then building the result based on 'order'. It handles remaining characters well.
2. Time Complexity: Correctly identified as O(N) where N is the length of 's'. The order loop is O(26) which is constant.
3. Space Complexity: Correctly identified as O(1) since the hash map size is limited to 26 characters.
4. Code Quality: Good structure, but could improve with:
   - More descriptive variable names (e.g., 'countMap' instead of 'hashMap')
   - Consistent indentation
   - Comments explaining the logic steps
5. Efficiency: The solution is efficient, but could potentially use an array of size 26 instead of a Map for slightly better performance.

Problem 2: Longest Substring Without Repeating Characters
1. Correctness: The solution has a logical error. When a duplicate is found, it moves 'i' but doesn't clear characters before the new 'i' from the map, which could lead to incorrect calculations.
2. Time Complexity: Claimed O(n) is correct for a proper implementation, but current implementation might not achieve this due to the bug.
3. Space Complexity: Correctly identified as O(1) (limited to character set size).
4. Code Quality: 
   - Good variable naming
   - Could use more comments
   - Initializing maxNo to -Infinity is unnecessary since 0 would suffice
5. Efficiency: Needs to fix the bug to achieve proper efficiency. After fixing, it would be optimal.

General improvements needed:
- Add more comments explaining the approach
- Handle edge cases explicitly (empty strings, etc.)
- For Problem 2, fix the logical error by clearing the map when moving 'i'
- Consider using more semantic variable names

super30admin:
Let's evaluate each problem separately.

Problem 1: Custom Sort String
1. Correctness: The solution correctly implements the custom sort by first counting characters in 's' and then building the result based on 'order'. It handles remaining characters appropriately.
2. Time Complexity: Correctly identified as O(N) where N is length of 's'. The order loop is O(26) which is constant.
3. Space Complexity: Correctly identified as O(1) since the hash map size is limited to 26 characters.
4. Code Quality: Good structure, but could improve with:
   - More descriptive variable names (e.g., 'countMap' instead of 'hashMap')
   - Consistent spacing (some lines have extra spaces)
   - Could use Array.from() with fill() for potentially cleaner character repetition
5. Efficiency: The solution is already efficient, but could consider using a frequency array instead of Map since we know the domain is lowercase letters.

Problem 2: Longest Substring Without Repeating Characters
1. Correctness: The solution has a logical error. When a duplicate is found, it moves 'i' but doesn't clear characters before the new 'i' from the map, which could lead to incorrect calculations.
2. Time Complexity: Claimed O(n) is correct for a proper implementation, but current implementation may not achieve this due to the logical error.
3. Space Complexity: Correctly identified as O(1) since the map size is limited by character set.
4. Code Quality: Good structure, but needs:
   - Better handling of edge cases (empty string)
   - Clearer comments explaining the algorithm
   - The initial split() is unnecessary as strings are already iterable
5. Efficiency: Needs to fix the logical error to achieve proper efficiency. Should clear old characters from the map when moving 'i'.

PR #1122, opened 2025-07-06
Strings-1 Completed!

Hitesh1090:


super30admin:
1. Correctness:
   - For `customSortString`, the solution correctly sorts the string `s` based on the order specified in `order` and appends remaining characters at the end. This matches the expected behavior.
   - For `longestSubstringWithoutRepeats`, the solution uses a sliding window approach with a hashmap to track character counts, which is a standard and correct approach for this problem.

2. Time Complexity:
   - `customSortString`: The student correctly identifies the time complexity as O(n), where n is the length of `s`. This is accurate as each character is processed once.
   - `longestSubstringWithoutRepeats`: The student correctly identifies the time complexity as O(n), where n is the length of `s`. This is accurate as each character is processed at most twice (once by `r` and once by `l`).

3. Space Complexity:
   - `customSortString`: The student claims O(1) auxiliary space, which is misleading. The space complexity is O(k), where k is the number of unique characters in `s`, as the hashmap stores counts for each unique character.
   - `longestSubstringWithoutRepeats`: The student correctly identifies the space complexity as O(min(m, n)), where m is the size of the character set and n is the length of the string. This is accurate.

4. Code Quality:
   - The code is generally well-structured and readable. Variable names are descriptive (`hmap`, `res`, `mlength`, etc.).
   - For `customSortString`, the use of `i*j` to concatenate characters is concise but could be clarified with a comment for readability.
   - For `longestSubstringWithoutRepeats`, the sliding window logic is clear, but the inner `while` loop could be optimized further by directly moving `l` to the position after the last occurrence of `s[r]` instead of incrementing one by one.

5. Efficiency:
   - Both solutions are efficient for their respective problems. However, for `longestSubstringWithoutRepeats`, the inner `while` loop could be optimized by storing the last occurrence of each character and jumping `l` directly to `max(l, last_occurrence[s[r]] + 1)`.

PR #1125, opened 2025-07-07
LC complete

asmodaks30:


super30admin:
Let's evaluate each solution separately:

1. LC3.cpp (Longest Substring Without Repeating Characters):
- Correctness: The solution correctly implements the sliding window approach with a hash map to track character positions. It handles duplicates by adjusting the left pointer.
- Time Complexity: O(n) - Correct, as each character is processed once.
- Space Complexity: O(1) - Technically correct if we assume a fixed character set (ASCII), but could be O(min(m,n)) where m is character set size.
- Code Quality: Good variable naming and structure. Could benefit from comments explaining the logic.
- Efficiency: Well optimized. One minor improvement could be to break early if remaining characters can't beat current max.

Potential edge cases: Empty string, all unique characters, all same characters.

2. LC791.cpp (Custom Sort String):
- Correctness: Correctly implements the counting sort approach for custom ordering.
- Time Complexity: O(m+n) - Correct, where m is order length and n is s length.
- Space Complexity: O(1) - Correct, using fixed-size array for counts.
- Code Quality: Clear and readable. Good use of stringstream. Could add comments about the two-pass approach.
- Efficiency: Well optimized. No significant improvements needed.

General strengths:
- Both solutions use appropriate data structures and algorithms
- Good time/space complexity analysis
- Clean, readable code

Areas for improvement:
- Could add more comments explaining the logic
- Could handle edge cases more explicitly (though current implementations should handle them)
- For LC3.cpp, consider adding early termination condition

super30admin:
1. Correctness:
   - LC3.cpp: The solution correctly implements the sliding window approach to find the longest substring without repeating characters. It handles edge cases well, such as empty strings or strings with all unique characters.
   - LC791.cpp: The solution correctly implements the custom sort based on the given order. It handles characters not present in the order string by appending them at the end.

2. Time Complexity:
   - LC3.cpp: The time complexity is O(n), as each character is processed exactly once.
   - LC791.cpp: The time complexity is O(m + n), where m is the length of the order string and n is the length of the string s. This is optimal for this problem.

3. Space Complexity:
   - LC3.cpp: The space complexity is O(min(m, n)), where m is the size of the character set (256 for ASCII) and n is the length of the string. The student incorrectly stated O(1), which is only true if the character set is fixed and small (e.g., ASCII).
   - LC791.cpp: The space complexity is O(1) because the size of the vector is fixed (26 for lowercase English letters).

4. Code Quality:
   - LC3.cpp: The code is well-structured and readable. Variable names are meaningful. The use of an unordered_map to track character positions is appropriate. However, the comment about space complexity could be more accurate.
   - LC791.cpp: The code is clean and efficient. The use of a vector to count character occurrences is a good choice. The stringstream is used effectively to build the result.

5. Efficiency:
   - LC3.cpp: The solution is efficient, but the space complexity comment should be corrected.
   - LC791.cpp: The solution is optimal for this problem. No further optimizations are needed.

PR #1126, opened 2025-07-07
Done Strings-1

BhavdeepSingh15:


super30admin:
1. Correctness:
   - For `customSortString`, the solution correctly implements the custom sorting based on the given order. It handles characters not in the order string by appending them at the end, which is correct.
   - For `lengthOfLongestSubstring`, the solution correctly implements the sliding window approach to find the longest substring without repeating characters.

2. Time Complexity:
   - `customSortString`: O(n + m), where n is the length of `s` and m is the length of `order`. This is efficient.
   - `lengthOfLongestSubstring`: O(n), where n is the length of `s`. This is optimal for this problem.

3. Space Complexity:
   - `customSortString`: O(n) for the HashMap and StringBuilder, which is reasonable.
   - `lengthOfLongestSubstring`: O(min(m, n)), where m is the size of the character set. This is optimal.

4. Code Quality:
   - The code is generally well-structured and readable.
   - For `customSortString`, the variable names are clear, but the loop for appending characters could be simplified using `String.repeat` (if using Java 11+) or a more concise loop.
   - For `lengthOfLongestSubstring`, the code is clean and follows best practices.

5. Efficiency:
   - Both solutions are efficient and follow standard approaches.
   - No major optimizations are needed, but minor readability improvements could be made, such as using helper methods or more concise loops where applicable.

PR #1129, opened 2025-07-17
solved strings problem using hashmap

IshaaS:


super30admin:
Strengths:
- The student demonstrates a good understanding of the problems and provides efficient solutions.
- The use of HashMap and HashSet is appropriate and well-implemented.
- The student provides multiple approaches for Problem2, showing versatility.

Areas for Improvement:
- In Problem1, the second loop over map.keySet() could be simplified by directly iterating over the remaining characters in the HashMap without the null check, as the HashMap will not contain null values.
- In Problem2, the HashSet approach could be slightly optimized by removing the redundant set.remove(s.charAt(end)) before adding it back in the next line.

super30admin:
Strengths:
- The student has provided efficient solutions with optimal time and space complexity.
- The student has implemented multiple approaches for Problem2, showing a good understanding of the problem.

Areas for Improvement:
- In Problem1, the redundant check `if(map.get(c)!=null)` inside the loop for remaining characters can be removed since the loop is already iterating over `map.keySet()`.
- The comments for Problem2 could be more detailed to explain why the hashmap approach is more efficient in certain cases.

PR #1130, opened 2025-07-19
Completed Strings-1

hrandhi-5303:


super30admin:
Strengths:
- The use of Counter is appropriate and simplifies the counting logic.

Areas for Improvement:
- Consider using a list comprehension for the final concatenation to make the code slightly more concise.
- The variable names could be more descriptive (e.g., `count` could be `char_count`).
- The second file (`longestSubString.py`) is unrelated to the problem and should not be included in the evaluation.

PR #1131, opened 2025-08-06
Strings 1 done

GuptaDivij:


super30admin:
- The student's solution is well-thought-out and efficient. The approach of using a hashmap to count frequencies and then building the string based on the order is clever and avoids the overhead of sorting.
- The comments are helpful and explain the approach clearly.
- The code is clean and follows good practices, such as using StringBuilder for efficient string concatenation.
- One minor suggestion would be to consider using an array of size 26 for counting frequencies instead of a hashmap, as the problem deals with lowercase English letters, which could potentially be slightly more efficient. However, the current approach is perfectly acceptable.

PR #1136, opened 2025-08-18
strings-1 done

NUPUR0307:


super30admin:
Strengths:
- The solutions are correct and handle edge cases well.
- The approach for Problem1 is more efficient than the reference solution by avoiding sorting.
- Code is clean, readable, and well-commented.

Areas for Improvement:
- The space complexity claim for Problem1 should be corrected to O(k) where k is the number of unique characters (up to 26), not O(1).
- For Problem2, while the solution is correct, it might be worth mentioning that the space complexity is O(min(m, n)) where m is the size of the character set (256 for ASCII) and n is the string length, as the hashset can grow up to the size of the character set.

PR #1138, opened 2025-08-26
problem 1 and 2 added

Tejas274:


super30admin:
Let me evaluate each solution separately.

For problem1-791-customsortstring.py:
1. Correctness: The solution correctly implements the custom sort by first counting character frequencies, then building the result in the order specified, and finally appending remaining characters. This approach is correct.
2. Time Complexity: O(n + m), where n is length of s and m is length of order. This is optimal.
3. Space Complexity: O(n) for storing character frequencies, which is reasonable.
4. Code Quality: The code is readable but could be improved by:
   - Using collections.defaultdict for cleaner frequency counting
   - Using more descriptive variable names (item_freq could be char_count)
   - The nested loops could be simplified using string multiplication
5. Efficiency: The solution is already efficient, but could be made more Pythonic.

For problem2-3-longestsubstringworepetingcharacters.py:
1. Correctness: The sliding window approach is correct, but there's a logical error in the else condition where it updates map_letter[s[j]] unnecessarily. The print statement should be removed.
2. Time Complexity: O(n) which is optimal for this problem.
3. Space Complexity: O(min(m,n)) where m is character set size, which is optimal.
4. Code Quality: Could be improved by:
   - Removing debug print statement
   - Adding comments to explain the sliding window logic
   - Simplifying the conditional logic which is currently a bit convoluted
5. Efficiency: The solution is mostly efficient but the logic could be streamlined by removing redundant updates to map_letter.

Common improvements needed:
- Add docstrings explaining the function purpose and parameters
- Consider using more Pythonic constructs (like defaultdict, Counter)
- Add edge case handling comments (empty strings, etc.)
- Consistent spacing around operators

PR #1139, opened 2025-08-26
Completed Strings-1

Paneripatel:


super30admin:
Let's evaluate each file separately.

1.py (Custom Sort String)
1. Correctness: The solution correctly implements the custom sort by first counting characters in 's' and then appending them in the order specified by 'order'. Any remaining characters are appended at the end. This approach is correct.
2. Time Complexity: O(m + n) where m is length of order and n is length of s. This is optimal.
3. Space Complexity: O(n) for the hashMap and result list, which is reasonable.
4. Code Quality: The code is well-structured but could be improved:
   - Use defaultdict from collections to simplify counting
   - The nested loops could be replaced with string multiplication (e.g., c * count)
   - Variable names could be more descriptive (e.g., 'count' instead of 'times')
5. Efficiency: The solution is already efficient, but could be made more Pythonic.

2.py (Longest Substring Without Repeating Characters)
1. Correctness: The sliding window approach with hashMap is correctly implemented to track character positions and adjust the window. This is a standard optimal solution.
2. Time Complexity: O(n) where n is length of s, which is optimal.
3. Space Complexity: O(min(m, n)) where m is character set size, which is optimal.
4. Code Quality: The code is clean and efficient:
   - Good use of sliding window technique
   - Clear variable naming
   - Proper handling of edge cases
5. Efficiency: The solution is optimal and well-implemented.

General suggestions:
1. Both solutions could benefit from docstrings explaining the approach
2. Consider adding type hints for return values
3. Add comments for complex logic (especially in the sliding window solution)
4. The initial null check in File 2 could be simplified to `if not s:`

PR #1140, opened 2025-08-28
solved 2 questions

nsh2207:


super30admin:
Strengths:
- The student has demonstrated a good understanding of the problems and has provided efficient solutions.
- The use of hash maps and sets is appropriate and well-implemented.

Areas for Improvement:
- For customSortString, while the solution is correct and efficient, it might be worth mentioning that the approach assumes a fixed character set (lowercase letters). If the problem were extended to include other characters, the solution would need adjustment.
- For lengthOfLongestSubstring, the while loop condition 'l <= n-1' could be written more clearly as 'l < n' for better readability.

PR #1145, opened 2025-09-22
Completed Strings-1

AneerudhS27:


super30admin:
Strengths:
- The customSortString solution is efficient and uses a clever approach with a hashmap.
- The space complexity is optimal for both solutions.

Areas for Improvement:
- The lengthOfLongestSubstring solution has a critical error in the loop bounds (iterating up to 256 instead of s.length()). This needs to be fixed to avoid out-of-bounds access and incorrect results.
- The function name customSorting should match the problem name customSortString for consistency.
- Consider adding input validation or edge case handling (e.g., empty strings) for robustness.

PR #1151, opened 2025-10-07
Completed String-1, 2 problems

meetrpatels30:


super30admin:
Strengths:
- The student has implemented both solutions correctly and efficiently.
- The time and space complexity analyses are accurate.
- The code is clean, readable, and follows best practices.

Areas for Improvement:
- For `customSortString`, while the solution is correct, it might be worth mentioning that the space complexity is O(1) only if the alphabet size is fixed (e.g., lowercase English letters). For a general case, it would be O(k), where k is the number of unique characters in `s`.
- In `lengthOfLongestSubstring`, the comment "Space complexity - O(1)" could be clarified to note that it assumes a fixed alphabet size.

PR #1152, opened 2025-10-07
Strings-1 completed

sathvikareddy18:


super30admin:
Strengths:
- The use of a frequency map is a good approach and avoids the need for sorting.
- The code is clean, readable, and follows Pythonic practices.

Areas for Improvement:
- Consider using a list comprehension for the final loop where characters not in `order` are appended. For example: `sb.extend([c] * cnt for c, cnt in smap.items())`.
- The solution for `lengthOfLongestSubstring` is not part of the problem being evaluated, but it is correct and efficient. Ensure that only the relevant solution is submitted when asked for a specific problem.

PR #1156, opened 2025-10-12
Implemented Strings-1

johar-inderpreet:
Implemented Strings-1


super30admin:
Strengths:
- The student has implemented both solutions correctly and efficiently.
- The use of frequency maps and sliding window techniques demonstrates a good understanding of the problems.
- Code is clean, readable, and well-commented.

Areas for Improvement:
- For `CustomSortString`, while the solution is correct, it might be worth mentioning that the problem could also be solved using sorting (as in the reference solution) for comparison, even though the student's approach is more efficient.
- In `LongestSubstring`, the variable `length` is defined but not used, which could be removed to avoid confusion.

PR #1157, opened 2025-10-12
String-1 Completed

Mike9330:


super30admin:
- Strengths:
  - The student demonstrates a strong understanding of time and space complexity, providing solutions that are more efficient than the reference in some cases.
  - The code is clean, well-commented, and easy to understand.
  - The use of a frequency map in `customSortString` is a clever optimization that avoids the need for sorting.
- Areas for Improvement:
  - While the solution for `customSortString` is efficient, it might be worth mentioning that the space complexity is technically O(1) only if we consider the number of unique characters as constant (which is reasonable for this problem).
  - In `lengthOfLongestSubstring`, the use of a fixed-size array for ASCII characters is efficient, but it might be worth noting that this approach may not generalize to Unicode characters without modification.

PR #1158, opened 2025-10-13
Completed Strings 1

Saideepthi123:


super30admin:
- Strengths:
  - Both solutions are correct and efficient, matching the expected time and space complexities.
  - Good use of Pythonic constructs like list comprehensions and lambda functions.

- Areas for Improvement:
  - In `customSortString`, the initial comment about time complexity could be clearer. The term "O(nlogn) + O(m)" is correct but could be simplified to "O(n log n)" since O(m) is dominated by O(n log n) when m is the length of `order` and n is the length of `s`.
  - In `lengthOfLongestSubstring`, the comment about space complexity could explicitly mention that the fixed-size array (256) is considered constant space.

PR #1162, opened 2025-10-28
Completed Strings-1

nikhilsaimarri:


super30admin:
Strengths:
- The solution is efficient with optimal time and space complexity.
- The approach of using a hashmap for frequency counting and then constructing the string is elegant.

Areas for Improvement:
- While the solution is correct and efficient, it might be worth mentioning in comments why the space complexity is O(1) (due to the fixed-size alphabet).
- The solution for "Longest Substring without Repeat Characters" is included but not part of the evaluation since it wasn't part of the reference problem. It would be better to focus on one problem at a time for clarity.

PR #1163, opened 2025-11-04
Strings 1

v-hemish:


super30admin:
- Strengths:
  - Both solutions are correct and efficient, with optimal time and space complexity.
  - The use of a frequency map in problem1.py is a clever approach that avoids sorting.
  - The sliding window technique in problem2.py is well-implemented and handles edge cases correctly.

- Areas for Improvement:
  - Add docstrings to explain the purpose and logic of each method.
  - Use more descriptive variable names to enhance readability (e.g., 'mp' could be 'char_count', 'sett' could be 'char_set').
  - In problem1.py, consider using a list comprehension for the final loop to append remaining characters, which could make the code more concise.

PR #1164, opened 2025-11-12
Completed Strings-1

nitishred13:


super30admin:
- **Strengths**: 
  - The `customSortString` solution is well-thought-out and efficient, using a HashMap to count characters and then constructing the result string in the desired order.
  - The code is readable and follows good practices, such as using `StringBuilder` for string concatenation.
- **Areas for Improvement**:
  - The `LongestSubString` solution does not correctly implement the sliding window technique. The logic for updating the window and tracking the maximum length is flawed. Consider reworking this solution to correctly handle the sliding window and avoid nested loops.
  - The variable names in `LongestSubString` could be more descriptive (e.g., `temp` could be `currentSubstring`).
  - The solution for `LongestSubString` should handle edge cases more gracefully, such as empty strings or strings with all unique characters.

PR #1167, opened 2025-11-20
Strings-1

PrasiddhShah:


super30admin:
Strengths:
- Good problem-solving approach for both problems.
- Correct implementation of efficient algorithms.
- Clean code structure and good variable naming.
- Appropriate use of data structures (HashMap, HashSet).

Areas for Improvement:
- Be more precise about space complexity analysis (especially when character set is limited).
- Consider using array instead of HashMap for problem1.java since character set is limited to lowercase English letters.
- Comments could be more detailed about why certain approaches were chosen.
- For problem2.java, the space complexity could be clarified as O(min(m, n)) where m is character set size.

PR #1168, opened 2025-12-16
Strings-1 Solutions

imran514:


super30admin:
Strengths:
- The student demonstrates a good understanding of different approaches to solving the problem, from naive to optimized solutions.
- The code is well-commented and easy to follow, with clear explanations of the algorithm approach.
- The student shows the ability to optimize solutions, as seen in the transition from the naive approach to the counting approach in the "Custom Sort String" problem.

Areas for Improvement:
- For the "Custom Sort String" problem, the first solution could be improved by using a counting approach from the start, as the naive approach is less efficient.
- In the "Longest Substring Without Repeating Characters" solutions, the student could consider adding more edge case tests (e.g., empty string, all unique characters) to ensure robustness.
- The student could further optimize the "Custom Sort String" problem by using an array instead of a HashMap for counting, as the problem is constrained to lowercase English letters.

PR #1169, opened 2026-01-03
Completed Strings-1

addi091:


super30admin:
Strengths:
- Excellent job on both solutions! Your implementations are not only correct but also more efficient than the reference solutions in terms of both time and space complexity for CustomSortString.
- Good use of HashMap to track character frequencies and indices.
- Clear and helpful comments explaining the approach.
- Proper variable naming and code organization.

Areas for Improvement:
- For CustomSortString, while your solution is more efficient, it might be worth mentioning that the O(1) space complexity assumes a fixed character set (like lowercase English letters).
- Consider adding some edge case testing in comments (like empty strings or strings with all unique characters for LengthOfLongestSubstring).

PR #1170, opened 2026-01-03
done

PurvaT-11:


super30admin:
- Strengths:
  - The time and space complexity are optimal.
- Areas for Improvement:
  - The comment at the top of the file mentions a sliding window approach, which is unrelated to this problem. This could be confusing and should be removed or corrected.
  - The solution could benefit from a brief explanation of the approach in the docstring, similar to the comment provided.

PR #1171, opened 2026-01-03
Strings-1 complete

rvuyyuru7:


super30admin:
Strengths:
- The student demonstrates a good understanding of the problems and provides efficient solutions.
- The use of frequency maps and sliding window techniques is appropriate and well-implemented.

Areas for Improvement:
- For "Custom Sort String", the student could consider using an array instead of a `HashMap` for the frequency map since the problem is constrained to lowercase English letters (26 characters). This would slightly improve performance.
- In "Longest Substring Without Repeating Characters", the initial comment mentions "256 ASCII characters," but the problem constraints might allow for a smaller array (e.g., 128 for standard ASCII). Clarifying this in the comment would be helpful.

PR #1173, opened 2026-01-05
Solved String-1 problems

aahiawasthee:


super30admin:
Strengths:
- The solution is correct and efficiently implemented.
- The use of a hashmap to count frequencies is a good approach.

Areas for Improvement:
- The space complexity claim should be corrected to O(k) where k is the number of unique characters in s.
- The solution for `longest-substring.java` was not requested but was included. It's better to focus on the problem at hand unless multiple solutions are required.

PR #1174, opened 2026-01-26
Completed Strings-1

pranavs0133:


super30admin:
Strengths:
- For Problem1, you implemented an efficient O(n) solution that is faster than the sorting-based reference solution.
- You correctly handled the case for characters not in `order` by appending them at the end.
- For Problem2, you used an optimal sliding window approach with a hash map to track indices, which is a standard and efficient solution.

Areas for Improvement:
- In Problem1, while the solution is efficient, it uses a hash map which might have a constant overhead. However, this is negligible given the constraints.
- Consider using an integer array of size 26 for frequency counting in Problem1, since the input is limited to lowercase letters. This could be slightly more efficient in terms of both time and space (avoiding hash map overhead).
- In Problem2, the initial value of `max` is set to `Integer.MIN_VALUE`, but if the string is empty, it returns 0 correctly. However, it might be clearer to initialize `max` to 0 and avoid the ternary operator at the end.

Overall, both solutions are correct and efficient. The code quality is high.

PR #1175, opened 2026-02-10
Custom Sort String solution

dbkuppagiri:


super30admin:
Strengths:
- The solution is correct and efficiently handles the problem.
- The approach is intuitive and well-explained in the comments.
- The use of a Map for frequency counting is appropriate and efficient.

Areas for Improvement:
- Consider using an array for building the result instead of string concatenation in loops for better performance in larger inputs. For example, you could push characters to an array and then join them at the end.
- Variable naming: `charFreq` is clear, but using `count` might be simpler. Also, consider using more descriptive names for loops (e.g., `for (const char of s)` is good, but the inner loop variable `i` is standard).
- Although not necessary, you could precompute the result length to avoid dynamic resizing if using an array.

Example improved code:
```javascript
var customSortString = function (order, s) {
    const freq = new Map();
    for (let char of s) {
        freq.set(char, (freq.get(char) || 0) + 1);
    }
    let result = [];
    for (let char of order) {
        if (freq.has(char)) {
            result.push(char.repeat(freq.get(char)));
            freq.delete(char);
        }
    }
    for (let [char, count] of freq) {
        result.push(char.repeat(count));
    }
    return result.join('');
};
```
Note: Using `repeat` is concise, but if performance is critical (though not necessary here), building the string without `repeat` might be slightly faster.

super30admin:
Strengths:
- The solution is efficient with linear time complexity, which is better than the reference solution's sorting approach.
- The use of a frequency map is appropriate and efficiently implemented.

Areas for Improvement:
- While the solution is efficient, the final loop for remaining characters could be optimized by using a for-of loop directly on the Map entries, which is already done. However, note that the order of remaining characters is arbitrary, which is acceptable.
- Consider using a StringBuilder-like approach for building the result string to avoid potential performance issues with string concatenation in JavaScript. In JavaScript, using an array to collect characters and then joining them at the end might be more efficient for large strings. For example:
  let resArray = [];
  // instead of res += char, do resArray.push(char);
  return resArray.join('');
  This is because string concatenation in a loop can be O(n^2) in worst-case due to immutable strings.
- The variable name `charFreq` is clear, but simply `count` might be sufficient.

Overall, the solution is excellent and outperforms the reference solution in terms of time complexity.

PR #1176, opened 2026-02-10
Length Of Longest Substring Solution

dbkuppagiri:


super30admin:
It seems there might have been a mix-up in the problems. The solution you provided is for "Longest Substring Without Repeating Characters", but the problem you were asked to solve is "Custom Sort String". 

For the Custom Sort String problem, you need to rearrange the characters in string `s` according to the order defined by string `order`. Characters in `s` that are not in `order` can be placed anywhere.

A common approach is to:
1. Create a mapping from each character to its priority (index in `order`). For characters not in `order`, you can assign a default value (like the length of `order` or beyond) so they are placed after the ordered characters.
2. Convert `s` into an array of characters and sort them based on the priority mapping.
3. Join the sorted array back into a string.

In JavaScript, you could implement it like this:

```javascript
var customSortString = function(order, s) {
    const orderMap = new Map();
    for (let i = 0; i < order.length; i++) {
        orderMap.set(order[i], i);
    }
    return s.split('').sort((a, b) => {
        const aIndex = orderMap.has(a) ? orderMap.get(a) : -1;
        const bIndex = orderMap.has(b) ? orderMap.get(b) : -1;
        return aIndex - bIndex;
    }).join('');
};
```

Alternatively, you can use an array of size 26 to store the priorities for efficiency.

Please review the problem statement again and implement the correct solution. Your current solution is for a different problem.

dbkuppagiri:
There are two problems in the repo, Am i missing something?

super30admin:
It appears that you have submitted a solution for the "Longest Substring Without Repeating Characters" problem instead of the "Custom Sort String" problem. Please review the problem statement carefully. For the Custom Sort String problem, you need to rearrange the characters in string `s` according to the custom order defined by string `order`. 

To solve the Custom Sort String problem, you can consider the following approach:
1. Create a mapping from each character in `order` to its index (or priority).
2. For characters not in `order`, you can assign a default value (like the length of `order` or beyond) so they are placed at the end.
3. Convert string `s` into an array of characters and sort them based on the mapping you created.
4. Join the sorted array back into a string and return it.

Here is a JavaScript example similar to the reference solution in Java:

```javascript
var customSortString = function(order, s) {
    const orderMap = new Map();
    for (let i = 0; i < order.length; i++) {
        orderMap.set(order[i], i);
    }
    
    const arr = s.split('');
    arr.sort((a, b) => {
        const aIndex = orderMap.has(a) ? orderMap.get(a) : order.length;
        const bIndex = orderMap.has(b) ? orderMap.get(b) : order.length;
        return aIndex - bIndex;
    });
    
    return arr.join('');
};
```

This solution has a time complexity of O(n log n) due to sorting, and space complexity of O(n) for storing the array and the map. Note that the constraints are small (s.length <= 200), so this is efficient enough.

Please make sure to submit the correct solution for the given problem.

PR #1177, opened 2026-02-10
Completed Strings-1 problems

thondaha:
This PR includes two problems in Strings -1 repo
1. custom-sort-string - https://leetcode.com/problems/custom-sort-string/description/
2. longest-substring-without-repeating-characters - https://leetcode.com/problems/longest-substring-without-repeating-characters/description/
It also includes a test class to verify the above algorithms are working as expected or not.

### DIFF
--- a/Test/TestStrings1Problems.java
+++ b/Test/TestStrings1Problems.java
@@ -1,0 +1,16 @@
+public class TestStrings1Problems {
+    public static void main(String[] args) {
+        // Testing Custom Sorting of the String
+        CustomSortString cs = new CustomSortString();
+        String order = "cba";
+        String s = "abcd";
+        System.out.println(cs.customSortString(order, s));
+        String order1 = "bcafg";
+        String s1 = "abcd";
+        System.out.println(cs.customSortString(order1, s1));
+        //Testing Longest Substring without Repeat Characters
+        LongestSubString ls = new LongestSubString();
+        String l = "abcabcbb";
+        System.out.println("The length of the longest substring is " + ls.lengthOfLongestSubstring(l) );
+    }
+}
--- a/src/CustomSortString.java
+++ b/src/CustomSortString.java
@@ -1,0 +1,40 @@
+/*
+problem - 791. Custom Sort String
+Approach - First we count how many times each character appears in s using a map.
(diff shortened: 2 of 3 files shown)


super30admin:
Strengths:
- The solutions for both problems are correct and efficient.
- The code is well-commented, explaining the approach and complexity.
- The use of HashMap for frequency counting in the Custom Sort String problem is intuitive, and the sliding window for the substring problem is appropriately implemented.

Areas for Improvement:
- For the Custom Sort String problem, consider using an integer array of size 26 for frequency counting instead of a HashMap. This would be more efficient in terms of constant factors and is common for problems with fixed character sets.
- In the Custom Sort String solution, when appending leftover characters, you can avoid iterating over the keySet by iterating over all 26 letters and checking if the frequency is non-zero. This ensures that the order of leftover characters is consistent (though the problem does not require a specific order for leftovers). Alternatively, you can iterate over the original string `s` and append characters not in `order`, but that would require an additional set for `order` characters to check membership quickly.
- For the Longest Substring solution, the inner while loop can be optimized by storing the index of each character's last occurrence. This would allow moving the slow pointer directly to the next position without a while loop, achieving a strict O(n) time complexity. However, the current solution is acceptable.

Overall, the solutions are well-implemented. Keep up the good work!

super30admin:
Strengths:
- The solution is correct and efficient, with a linear time complexity that beats the reference solution's sorting approach.
- The use of a HashMap to count frequencies is appropriate, and the handling of characters not in `order` is straightforward.

Areas for Improvement:
- Although the solution is efficient, the removal of keys from the map after processing the `order` string is not necessary. Instead, you could simply iterate over the remaining keys, but note that the keys not in `order` are appended at the end. However, the current approach is acceptable.
- The solution includes an unrelated class (LongestSubString) and a test class. For clarity and focus, it's better to submit only the required solution for the problem being evaluated.
- Consider using an array of size 26 for frequency counting instead of a HashMap, since the problem states that the strings consist of lowercase English letters. This would be slightly more efficient and use constant space as well.

Overall, the solution for Custom Sort String is excellent. Just ensure that future submissions are focused on the problem at hand.

PR #1178, opened 2026-02-10
String-1 HW : Length of Longest Substring

Amaan-29:


super30admin:
It seems you may have submitted the wrong files. The problem requires a solution for "Custom Sort String", but you provided a solution for "Longest Substring Without Repeating Characters". Please ensure you are working on the correct problem and submit the appropriate files.

For the Custom Sort String problem, you need to implement a method that sorts string `s` based on the custom order defined in string `order`. One common approach is to define a custom comparator that uses the order of characters in `order` and then sort the characters of `s` accordingly. Characters not in `order` can be placed arbitrarily.

Here's a suggested approach:
1. Create an array to store the priority of each character (based on their index in `order`).
2. Convert string `s` to a character array.
3. Sort the character array using a custom comparator that compares characters based on their priority from the array. For characters not in `order`, you can assign a default value (like the original character value or a fixed priority).
4. Convert the sorted array back to a string and return it.

Alternatively, you can use a counting approach to avoid sorting (which would be more efficient), but the sorting solution is acceptable given the constraints.

Please resubmit the correct solution for the Custom Sort String problem.

super30admin:
Strengths:
- The solution is correct and efficient.
- It uses a frequency map to count characters, which is a good approach.
- The code is readable and well-structured.

Areas for improvement:
- The frequency map initialization can be simplified using `getOrDefault` to avoid the if-check.
- The removal of keys is not strictly necessary; we could instead iterate over the frequency map and skip characters that are in `order` (but the current approach is fine).
- For a submission, the main method should be omitted unless required.

Final

PR #1179, opened 2026-02-10
2problems completed

BharathVuppala96:


super30admin:
Strengths:
- The solution is correct and efficient, with a linear time complexity which is better than the reference solution's O(n log n).
- The approach is straightforward: count frequencies, then output in the order specified.

Areas for Improvement:
1. Avoid using built-in names like `map` for variables. Instead, use descriptive names like `freq` or `char_count`.
2. You can simplify the frequency counting by using `collections.Counter` for better readability.
3. Instead of nested loops for appending characters, you can use string multiplication or extend the list with a list of repeated characters. For example: `word.extend([char] * freq[char])`.
4. Follow PEP8 guidelines: add spaces around operators and after commas. For example, `for i in range(len(s)):` should have a space after the comma if there are multiple parameters (but here it's fine), but also use more descriptive variable names than `i`, `j`, etc.
5. The solution for a different problem is included in the same file. Each problem should be in its own file or at least separated clearly. For this exercise, only the relevant solution should be provided.

Here's a revised version of your code that addresses these points:
```python

class Solution:
        freq = Counter(s)
        result = []
        
        for char in order:
            if char in freq:
                result.append(char * freq[char])
                del freq[char]
                
        for char, count in freq.items():
            result.append(char * count)
            
        return ''.join(result)
```

This version is more Pythonic and readable.

super30admin:
Your solution for "Custom Sort String" is well thought out and efficient. Here are some points to consider:

Strengths:
- You used a frequency counting approach which is optimal for this problem (O(n) time).
- The code is clear and easy to understand.

Areas for improvement:
- Avoid using `map` as a variable name because it is a built-in function in Python. Consider using `freq` or `frequency` instead.
- You can make the code slightly more concise by using `collections.Counter` to count the frequencies. For example:
  ```python
  freq = Counter(s)
  ```
- Instead of using nested loops to append characters multiple times, you can use:
  ```python
  word.extend([char] * freq[char])
  ```
  This would be more efficient and readable.
- Also, remember to handle the case when a character in `order` is not in `s` (your code already handles this correctly by checking `if order[i] in map`).

Overall, great job!

PR #1181, opened 2026-02-12
completed strings-1

anjanaouseph:


super30admin:
Your solution for "Custom Sort String" is well-implemented and efficient. You used a counting approach that has linear time complexity, which is better than the sorting-based reference solution. The code is clean and easy to understand.

Strengths:
- Correctness: Your solution correctly handles the problem requirements.
- Time Complexity: O(n) is optimal for this problem.
- Code Readability: Good variable names and comments.

Areas for improvement:
- The initial check for empty strings is not necessary due to the constraints, but it doesn't harm.
- Instead of using a while loop to decrement the count and append, you could do:
  ```python
  count = hashMap.get(ch, 0)
  result.append(ch * count)
  hashMap[ch] = 0   # to avoid later processing
  ```
  This might be slightly more efficient as it avoids multiple appends and multiple decrements. However, the current approach is acceptable.

Note: You also included a solution for a different problem (problem96.py) in your submission. Make sure to only submit the solution for the problem you are solving. For the purpose of this evaluation, we only considered the first file.

PR #1182, opened 2026-02-14
done Strings-1

sainathek1999:


super30admin:
Strengths:
- Both solutions are correct and efficient, meeting the problem requirements.
- The student uses appropriate data structures (HashMap, HashSet) and understands their properties.
- The time and space complexity analysis is accurate for both problems.

Areas for Improvement:
1. For problem1.java:
   - Instead of manually checking and updating the frequency map with `if-else`, use `fmap.put(c, fmap.getOrDefault(c, 0) + 1);` for conciseness.
   - When appending characters from the order, you can use a for loop with the count to append the character multiple times, which might be more readable than a while loop with a counter.
   - Consider using a StringBuilder for efficient string building, which you already did.

2. For problem2.java:
   - The inner while loop is correct, but you can optimize by storing the last occurrence index of each character in a map. This would allow you to jump the left pointer directly to `max(left, lastOccurrence + 1)` when a duplicate is found, avoiding the while loop. This approach is also O(n) but might be faster in practice.
   - The set removal in the inner loop is necessary, but with the index map approach, you can avoid explicit removals and just update the left pointer.

3. General:
   - Add comments to explain the logic steps, especially for problem2.java where the sliding window approach might benefit from clarification.
   - Use consistent variable names (e.g., `fmap` could be `freqMap` for clarity).

super30admin:
Strengths:
- For problem1, you implemented an efficient solution with linear time complexity, which is better than the reference solution that uses sorting.
- You used appropriate data structures (HashMap and StringBuilder) and provided clear comments.
- You correctly handled the removal of processed characters from the frequency map to avoid duplicates.

Areas for Improvement:
- In problem1, you can simplify the frequency map update using `fmap.put(c, fmap.getOrDefault(c, 0) + 1)` or `fmap.merge(c, 1, Integer::sum)` to make the code more concise.
- For problem2, the solution is incorrect. The sliding window approach requires careful management of the set to reflect the current window. When a duplicate is found, you should remove characters from the set starting from the left pointer until the duplicate character is removed, and then add the current character. However, your current code does not remove the duplicate character from the set. Consider this corrected approach:
  ```java
  Set<Character> set = new HashSet<>();
  int left = 0, max = 0;
      char c = s.charAt(right);
      while (set.contains(c)) {
          set.remove(s.charAt(left));
          left++;
      }
      set.add(c);
      max = Math.max(max, right - left + 1);
  }
  return max;
  ```
  This ensures the set always contains only the characters in the current window.

- Always test your solutions with edge cases, such as empty strings or strings with all duplicates, to catch logical errors.

PR #1183, opened 2026-02-14
Done Strings-1

vaishnavi2231:


super30admin:
Your solution for Custom Sort String is excellent! It is more efficient than the reference solution in terms of time complexity (O(m+n) vs O(n log n)) and is well-implemented. Here are some strengths and minor suggestions:

Strengths:
- You used a hashmap to count frequencies, which is efficient and straightforward.
- You correctly handled the order by iterating through `order` first and then appending the remaining characters.
- The code is clean and well-commented.

Suggestions:
- Although your solution is efficient, note that the problem constraints are small (s.length <= 200), so even an O(n log n) solution would be acceptable. However, your linear approach is optimal.
- You can use a list comprehension or a more concise way to build the result for the remaining characters. For example, instead of nested loops, you could do: `for char, count in hashmap.items(): res.extend([char] * count)`.
- Be cautious when submitting multiple solutions in one file. It seems you included a solution for another problem by mistake. Always ensure that each file contains only the solution for the intended problem.

Overall, great job! Your solution is correct and efficient.

super30admin:
Strengths:
- The solution is efficient with O(m + n) time complexity, which is better than the reference solution that uses sorting.
- The use of a hashmap to count frequencies is appropriate.

Areas for Improvement:
- While the solution is correct, it might be slightly less efficient in practice for very small strings due to the overhead of the hashmap. However, this is negligible.
- The solution uses `hashmap.keys()` to iterate over the remaining characters. Note that the order of these keys is arbitrary, but that is acceptable because the problem allows any order for characters not in `order`. So this is fine.
- Alternatively, you could avoid creating a list of keys by iterating directly over the hashmap, but the current way is acceptable.

Overall, the solution is excellent.

PR #1184, opened 2026-02-15
Strings-1 completed

sujaygijre:


super30admin:
Your solution for CustomSortString is correct and efficient. It has a linear time complexity which is better than the reference solution that uses sorting. The code is clean and easy to understand. Well done!

One minor suggestion: you don't need to explicitly erase the character from the map in the first loop. Since you have already decremented the count to zero, accessing `um[ch]` again would return 0 (because the map will create an entry with value 0 if it doesn't exist when using `operator[]`). However, by erasing, you avoid creating entries for characters that are not in `s` when you iterate over `order`. But note that if a character in `order` does not appear in `s`, then `um[ch]` would be created with value 0 when you access it in the while loop. Then you erase it. This is acceptable, but without erasing, it would still work because the while loop would not run for count 0. Alternatively, you could check if the character exists in the map before processing to avoid creating entries. However, your current approach is fine.

Another point: in the last loop, you are using a range-based for loop with `auto& [key, val]`. This is correct and modern C++. However, note that you are modifying `val` (by decrementing) and then appending the character. This is efficient.

Overall, great job!

super30admin:
Strengths:
- Your solution is efficient with linear time complexity, which is better than the reference solution that uses sorting.
- The approach of counting frequencies and then building the result in the order of `order` is straightforward and correct.
- You correctly handle the characters not in `order` by appending them at the end.

Areas for improvement:
- Consider using more descriptive variable names. For example, `um` could be renamed to `freq` or `charCount` to make the code more readable.
- In the first loop, you are counting the frequency of each character in `s`. This is correct, but note that when you iterate over `order`, you are using `um[ch]` which might insert a new key if `ch` is not present. However, since `order` contains unique characters and we are only accessing those that might be in `s`, it is safe. But to avoid unnecessary insertions, you could check if the character exists in the map before the while loop. Alternatively, you can rely on the fact that accessing a key that doesn't exist in an unordered_map (in C++) will create it with value 0, so the while loop condition will be false. This is acceptable.
- In the second loop, when you iterate over the remaining characters, you are using `um.erase(ch)` inside the first loop. This is not necessary because the while loop already reduces the count to zero. However, it doesn't harm. Alternatively, you could avoid erasing and just iterate over the map for the remaining characters.

Overall, your solution is excellent and more efficient than the reference solution. Well done!

PR #1185, opened 2026-02-15
Completed Strings-1

Nikesh46:


super30admin:
Strengths:
- The solution correctly implements the problem requirements.
- The time complexity is optimal (O(n + m)), which is better than the reference solution that uses sorting.
- The use of HashMap to count frequencies is appropriate and efficient.

Areas for Improvement:
- Code readability: Improve formatting by adding consistent spaces (e.g., `for(int i=0; i< s.length();i++)` should be `for (int i = 0; i < s.length(); i++)`).
- Avoid duplicate code: The while loops for appending characters can be abstracted into a helper method or replaced with a for-loop to make the code cleaner.
- Consider using an array of size 26 for frequency counting since the problem states that the strings consist of lowercase English letters. This would be more efficient in terms of constant factors and avoid the overhead of HashMap. For example:
  ```java
  int[] freq = new int[26];
  for (char c : s.toCharArray()) freq[c - 'a']++;
  ```
  Then, when iterating over `order`, you can get the frequency directly from the array. For the remaining characters, you can iterate over the array and append any character that has a frequency > 0 and is not in `order` (but note: you don't know which ones are in order? Actually, you can mark which ones are processed? Alternatively, you can first append all characters in the order, and then append all other characters by iterating over the frequency array. However, the current approach with HashMap is acceptable.

Example improved code using array:
```java
class Solution {
    public String customSortString(String order, String s) {
        int[] freq = new int[26];
        for (char c : s.toCharArray()) {
            freq[c - 'a']++;
        }
        StringBuilder res = new StringBuilder();
        for (char c : order.toCharArray()) {
            while (freq[c - 'a'] > 0) {
                res.append(c);
                freq[c - 'a']--;
            }
        }
        for (int i = 0; i < 26; i++) {
            while (freq[i] > 0) {
                res.append((char)('a' + i));
                freq[i]--;
            }
        }
        return res.toString();
    }
}
```

Overall, the student's solution is correct and efficient.

super30admin:
Strengths:
- The solution is correct and efficiently implemented with linear time complexity.
- The use of HashMap for frequency counting is appropriate.

Areas for Improvement:
- Instead of removing the character from the map after processing the order, you could simply not remove it and then in the final loop, only output characters that are not in the order. However, since the order string has unique characters and the map has at most 26 entries, the current approach is acceptable.
- Consider using an array of size 26 for frequency counting since the problem states that the strings consist of lowercase English letters. This would be more efficient in terms of constant factors and avoid the overhead of HashMap. For example:
  int[] freq = new int[26];
  for (char c : s.toCharArray()) freq[c - 'a']++;
  Then for each char in order, append the character repeated freq[c - 'a'] times and set freq[c - 'a'] to 0.
  Then iterate over the freq array to append the remaining characters.

Overall, the solution is very good. The only minor improvement would be to use an array instead of HashMap for better performance, but the current solution is correct and efficient.

PR #1186, opened 2026-02-28
Completed Strings-1

shreyargh:


super30admin:
Your solution for CustomSortString is correct and efficient. You used a frequency map to count characters and then built the result string by following the custom order. This approach has linear time complexity, which is better than the reference solution that uses sorting. Well done!

However, here are a few suggestions to improve:
1. Consider using an integer array of size 26 instead of a HashMap for counting frequencies. Since the problem constraints specify that all characters are lowercase English letters, an array would be more efficient and straightforward. For example:
   ```java
   int[] count = new int[26];
   for (char c : s.toCharArray()) count[c - 'a']++;
   ```
2. When appending characters from the order, you can directly use the array for frequency counts and avoid the HashMap operations.
3. For the leftover characters, you can iterate over the entire array (26 letters) and append any that have a count greater than zero. This avoids the need for an entry set iteration.
4. Please ensure that you only submit the solution for the problem being evaluated. The inclusion of another solution (LongestSubstringWithoutRepeatingCharacters) might be an error.

Here's a revised version of your code using an array:
```java
class Solution {
    public String customSortString(String order, String s) {
        int[] count = new int[26];
        for (char c : s.toCharArray()) {
            count[c - 'a']++;
        }
        
        StringBuilder sb = new StringBuilder();
        for (char c : order.toCharArray()) {
            while (count[c - 'a'] > 0) {
                sb.append(c);
                count[c - 'a']--;
            }
        }
        for (char c = 'a'; c <= 'z'; c++) {
            while (count[c - 'a'] > 0) {
                sb.append(c);
                count[c - 'a']--;
            }
        }
        return sb.toString();
    }
}
```

This version is more efficient and cleaner.

PR #1187, opened 2026-03-10
"Strings-1 done"

Sahithipsl470:


super30admin:
### Custom Sort String (Problem-1.py)
Your solution is well-structured and efficient. You correctly identified that using a counter to track character frequencies and then building the result in two passes (first for characters in `order`, then for the rest) is optimal. This approach avoids the overhead of sorting and achieves linear time complexity, which is better than the reference solution's O(n log n). 

One minor improvement: Instead of iterating over `count` (which is a dictionary) for the remaining characters, you can iterate over the original `s` or use a fixed set of characters to ensure deterministic behavior. However, since the problem allows any order for the remaining characters, your current approach is acceptable. Also, note that iterating over a dictionary keys might not be in a consistent order across different Python versions, but that doesn't affect correctness here.

Another point: You can use a list for `res` as you did, which is efficient for appending. Alternatively, you could use a list of strings and then join them at the end, which is what you did and is efficient in Python.

Overall, great job! Your solution is correct, efficient, and clean.

VERDICT: PASS

---

###  Longest Substring without Repeat Characters (Problem-2.py)
Strengths:
- The solution is efficient with O(n) time complexity, which is optimal for this problem.
- The code is concise and easy to understand, with clear variable names and comments.
- The use of enumerate for the right pointer and character is idiomatic in Python.

Areas for improvement:
- While the code is correct, it might be slightly inefficient in the worst-case scenario (all characters unique) because the inner while loop would not run, but in the worst-case scenario with all duplicates, it could run multiple times. However, overall it is efficient.
- Consider adding a check for an empty string at the beginning to avoid unnecessary operations, though it is not critical.
- The comments are good, but you could also explain why the sliding window approach is used and how it improves over brute force.

Overall, this is an excellent solution that meets the problem requirements efficiently.

VERDICT: PASS

PR #1188, opened 2026-03-15
solved

jeetusingh247:


super30admin:
### Custom Sort String (Custom_Sort.java)
Your solution is excellent! It correctly solves the problem with an optimal time complexity of O(n + m), which is better than the reference solution that uses sorting (O(n log n)). Your use of a frequency map and StringBuilder is efficient and clear.

Strengths:
- Correctness: Your solution produces the correct output by first appending characters in the order specified and then appending any remaining characters.
- Efficiency: The linear time complexity is optimal for this problem.
- Code Clarity: The code is well-commented and easy to follow.

Areas for Improvement:
- While your final loop over `map.keySet()` is correct, note that after processing the `order` string, the map contains only characters not in `order`. Since the map has at most 26 keys, it is efficient. However, you could consider using `map.entrySet()` to avoid multiple lookups for each character's count, but the difference is negligible.
- Pre-allocating the StringBuilder with `new StringBuilder(s.length())` might slightly improve performance by avoiding internal resizing, though it is not necessary.

Overall, this is a great solution. Keep up the good work!

VERDICT: PASS

---

###  Longest Substring without Repeat Characters (Longest_Substring.java)
Your solution is excellent! You have correctly implemented the sliding window technique with a HashSet, which efficiently solves the problem in O(n) time and O(1) space. Your code is clean and well-commented, making it easy to understand.

One suggestion for improvement: Instead of using a while loop to remove characters one by one until the duplicate is found, you can use a HashMap to store the last index of each character. This way, when you encounter a duplicate, you can immediately move the slow pointer to max(slow, lastOccurrence + 1). This optimization avoids the inner while loop and makes the algorithm slightly more efficient in practice. However, your current solution is still efficient and correct.

Here's an example of how you might implement that:
```java
    int n = s.length();
    int max = 0;
    Map<Character, Integer> map = new HashMap<>();
    for (int i = 0, slow = 0; i < n; i++) {
        char c = s.charAt(i);
        if (map.containsKey(c)) {
            slow = Math.max(slow, map.get(c) + 1);
        }
        map.put(c, i);
        max = Math.max(max, i - slow + 1);
    }
    return max;
}
```

Overall, great job! Your solution is optimal and well-written.

VERDICT: PASS

PR #1189, opened 2026-03-17
Strings 1 done

shukan26:


super30admin:
### Custom Sort String (CustomSortStrings.java)
Your solution is correct and efficient. You have used a frequency array to count characters and then built the result by first appending characters in the custom order and then appending the rest. This approach has a time complexity of O(n + m), which is better than the reference solution that uses sorting (O(n log n)). 

Strengths:
- You have achieved a linear time solution, which is optimal for this problem.
- You used a frequency array which is a common and effective technique.

Areas for improvement:
- The last loop can be optimized. Instead of iterating over `s` again and checking each character's frequency, you can iterate over the 26 letters in the frequency array. For each character that has a positive frequency after processing `order`, append it as many times as its frequency. This would be more efficient because it only checks 26 elements instead of n. Also, it avoids redundant checks for the same character multiple times in `s`.
- Currently, the last loop appends the remaining characters in the order they appear in `s`. This is acceptable per the problem, but if you iterate over the frequency array, the order of the appended characters might be different (alphabetical order). However, the problem does not require any specific order for the extra characters, so either way is fine. But if you want to preserve the original relative order of the extra characters (which is not required), you would need a different approach. However, the problem does not require that, so it's acceptable.

Here is a suggested improvement for the last part:
```java
// Instead of the last loop:
for (int i = 0; i < 26; i++) {
    while (freq[i] > 0) {
        result.append((char)('a' + i));
        freq[i]--;
    }
}
```
This loops over each of the 26 letters and appends any remaining characters. This is more efficient and avoids iterating over `s` again.

Overall, your solution is good and passes the requirements. With a small optimization, it can be made even better.

VERDICT: PASS

---

###  Longest Substring without Repeat Characters (LongestSubstringWithoutRepeatingCharacters.java)
Your solution demonstrates a good understanding of the sliding window technique, which is a common approach for this problem. However, there are a few issues that need to be addressed:

1. **Syntax Error**: You have defined the same class `LongestSubstringWithoutRepeatingCharacters` twice in the same file. This is not allowed in Java and will cause a compilation error. You should only have one class definition.

2. **IndexOutOfBoundsException Risk**: Although your outer while loop checks `right < s.length()`, the inner while loop uses `s.charAt(right)` without rechecking the bounds. However, since the inner while loop is inside the outer loop where `right` is guaranteed to be within bounds, this is safe. But note that the inner while loop might change `left` and `right` remains the same, so it's safe.

3. **Efficiency Improvement**: While your solution is O(n) in time, it can be optimized further. Currently, when a duplicate is found, you shrink the window from the left one character at a time. This can be improved by using a HashMap to store the last occurrence index of each character. Then, when a duplicate is found, you can move the left pointer directly to `max(left, lastOccurrence + 1)`. This avoids multiple unnecessary removals and makes the algorithm faster in practice.

4. **Code Clarity**: Your comments are helpful, but you should also consider adding more comments to explain why the inner while loop is necessary. For example, "We remove characters from the left until the duplicate character is removed from the set."

5. **Edge Cases**: Have you considered the case when the string is empty? Your code should handle it because `right < s.length()` will be false for an empty string, so it returns 0. This is correct.

Here is a revised version of your code without the duplicate class and with the HashMap optimization:

```java
import java.util.*;

public class LongestSubstringWithoutRepeatingCharacters {
        int left = 0;
        int max = 0;
        Map<Character, Integer> charIndexMap = new HashMap<>();
        
            char c = s.charAt(right);
            if (charIndexMap.containsKey(c)) {
                left = Math.max(left, charIndexMap.get(c) + 1);
            }
            charIndexMap.put(c, right);
            max = Math.max(max, right - left + 1);
        }
        return max;
    }
}
```

This version uses a HashMap to store the last index of each character. When a duplicate is found, it updates the left pointer to the next index after the last occurrence of the duplicate character. This approach is more efficient and avoids the inner while loop.

VERDICT: NEEDS_IMPROVEMENT

PR #1190, opened 2026-03-23
Strings-1

nagasai67:


super30admin:
### Custom Sort String (customSort.py)
Your solution is excellent! It efficiently solves the problem with a linear time complexity, which is better than the sorting-based reference solution. Here are some strengths and minor suggestions for improvement:

Strengths:
- You correctly identified the problem requirements and implemented a solution that meets them.
- The use of a frequency map is appropriate and efficient.
- The code is concise and easy to follow.

Areas for improvement:
- **String Concatenation**: Instead of using `res += ...` in loops, consider using a list to accumulate characters and then join them at the end. This avoids creating new string objects repeatedly. For example:
  ```python
  res_list = []
  for i in order:
      if i in cnt_mp:
          res_list.append(i * cnt_mp[i])
          del cnt_mp[i]
  for k, v in cnt_mp.items():
      res_list.append(k * v)
  return ''.join(res_list)
  ```
  This change would make the solution more efficient for large strings.

- **Variable Names**: While `cnt_mp` is understandable, using `freq_map` or `char_count` might be more descriptive.

- **Handling Missing Keys**: Your code checks `if i in cnt_mp` before accessing, which is good. Alternatively, you could use `collections.Counter` for `s` to simplify frequency counting. For example:
  ```python
  cnt_mp = Counter(s)
  ```
  This is more Pythonic and handles missing keys gracefully.

Overall, your solution is correct and efficient. With minor adjustments, it can be even better.

VERDICT: PASS

---

###  Longest Substring without Repeat Characters (longestSub.py)
Note: The verdict should be based on whether the student's solution is correct and efficient enough for the problem. For problems with multiple solutions, if the student's solution is correct but not the most optimal, it might still be acceptable if it meets the problem constraints. However, if the problem explicitly requires an optimal solution and the student's is not, then it should be marked as NEEDS_IMPROVEMENT.

Now, evaluate the student's solution.

VERDICT: NEEDS_IMPROVEMENT

PR #1191, opened 2026-03-24
Problem1 added

megharaykar:


super30admin:
### Custom Sort String (Problem1.py)
Your solution is well-thought and efficient. Here are some strengths and suggestions:

Strengths:
- You achieved a linear time complexity, which is better than the sorting approach in the reference solution.
- The code is straightforward and easy to follow.
- You correctly handled the characters not in `order` by appending them at the end.

Areas for Improvement:
- Consider using `collections.Counter` to simplify the frequency counting. For example:
  ```python
  freq = Counter(s)
  ```
  This would make the code more concise and idiomatic.
- The variable name `hm` (which stands for hash map) is acceptable, but `freq` or `freq_map` might be more descriptive.
- You can avoid the explicit loop for counting by using `Counter`. Also, when building the result, you can iterate over `order` and then the remaining keys, but note that `Counter` has a `keys` method which might be useful.

Example using Counter:
```python

    freq = Counter(s)
    res = []
    for ch in order:
        if ch in freq:
            res.append(ch * freq[ch])
            del freq[ch]
    for ch, count in freq.items():
        res.append(ch * count)
    return ''.join(res)
```
This version is more Pythonic and efficient in terms of code clarity.

VERDICT: PASS

---

###  Longest Substring without Repeat Characters
It seems there has been a mix-up in the problem you are solving. The code you provided is for "Custom Sort String", but the problem you are supposed to solve is "Longest Substring without Repeat Characters". Please double-check the problem statement and ensure you are working on the correct problem.

For "Longest Substring without Repeat Characters", you need to find the length of the longest substring without repeating characters. The reference solution uses a brute-force approach with two nested loops and a set to check for duplicates. However, there are more efficient approaches using a sliding window technique with O(n) time complexity.

Here are some tips for the correct problem:
- You can use a sliding window (two pointers) to traverse the string.
- Maintain a set or a dictionary to keep track of characters in the current window.
- When you encounter a duplicate, move the left pointer to remove characters until the duplicate is eliminated.
- Update the maximum length as you expand the window.

Example of an efficient solution in Python:
```
def lengthOfLongestSubstring(s: str) -> int:
    char_set = set()
    left = 0
    max_len = 0
    for right in range(len(s)):
        while s[right] in char_set:
            char_set.remove(s[left])
            left += 1
        char_set.add(s[right])
        max_len = max(max_len, right - left + 1)
    return max_len
```

Please implement the correct solution for the given problem.

VERDICT: NEEDS_IMPROVEMENT

PR #1192, opened 2026-03-24
Strings 1

shinjaneegupta:


super30admin:
### Custom Sort String (CustomSortString.py)
Your solution is excellent! You have correctly implemented an efficient approach that outperforms the sorting-based reference solution in terms of time complexity. Here are some strengths and minor suggestions:

Strengths:
- You used a frequency counter to efficiently track character counts, which is ideal for this problem.
- The approach is straightforward and easy to understand.
- You handled edge cases (characters not in `order`) correctly.

Areas for Improvement:
- While the code is clear, you can consider using a list comprehension or more concise methods for building the result. For example, instead of nested loops for appending characters, you could use string multiplication and concatenation, but the current approach is perfectly fine.
- You might want to avoid deleting keys from the frequency map during iteration over `order`, as it is not necessary. Instead, you can iterate over the keys in the frequency map at the end without deleting, but your method is correct and efficient.

Minor Note:
- The comment states "O(1)" for space complexity due to the constant-size alphabet, but the result list is O(n), so overall space is O(n). However, the constant factors for the frequency map are indeed O(1). This is a minor point and your understanding is correct.

Overall, great job!

VERDICT: PASS

---

###  Longest Substring without Repeat Characters (LongestSubstringWithoutRepeatingChars.py)
Your solution is excellent and demonstrates a good understanding of the sliding window technique. The code is efficient and handles the problem optimally. 

One minor improvement: The condition `slow <= index < fast` can be simplified to `slow <= index` because `index` (the last occurrence) is always less than `fast` since we haven't updated the map for the current `fast` yet. This makes the condition slightly more straightforward.

Also, consider adding a brief comment explaining why the space complexity is O(1) (due to the fixed size of the character set) to make it clear to readers.

Overall, great job!

VERDICT: PASS

PR #1193, opened 2026-03-26
Completed Strings-1

sarvanibaru:


super30admin:
### Custom Sort String (CustomSortString.java)
Your solution is efficient and correct. You have successfully implemented an approach that improves upon the reference solution's time complexity by avoiding sorting. Here are some points for improvement:

1. **Consistency in Output**: While the problem allows any order for characters not in `order`, using HashMap's keySet might lead to different outputs across runs due to the unordered nature. Consider iterating over the original string `s` for the remaining characters to ensure consistent output if needed, but note that the problem does not require this. Alternatively, you could use a LinkedHashMap to preserve insertion order, but it's not necessary.

2. **Efficiency in Iteration**: When appending characters from the map, you are iterating over the keySet and then appending each character multiple times. This is acceptable, but you can also iterate once over the entire keySet and append each character according to its frequency without nested loops by using a for-each loop and then appending the character repeated `frq` times. However, your current approach is clear and straightforward.

3. **Code Comments**: Your comments are good, but you could add more details about why you remove characters from the map after processing the `order` string. This helps clarify that you are only processing each character in `order` once and then handling the rest.

4. **Edge Cases**: Your solution handles edge cases well, such as when `s` has characters not in `order` and vice versa. However, consider mentioning how you handle duplicate characters in `s` (which you do correctly by counting frequencies).

Overall, your solution is excellent and demonstrates a good understanding of the problem. Keep up the good work!

VERDICT: PASS

---

###  Longest Substring without Repeat Characters (LongestSubstringWithoutRepeatingCharacters.java)
Your solution is excellent! You have implemented an optimal O(n) time complexity solution with O(1) space complexity. The code is clear and well-commented. Here are a few suggestions for improvement:

1. The comment about space complexity should be adjusted to account for the full range of characters mentioned in the problem (not just 26 alphabets). You could say: "O(1) because the set size is bounded by the number of distinct characters (which is at most 128 for ASCII, or a constant for any character set)."

2. Consider using a more efficient data structure for character tracking, such as a boolean array of size 128 (for ASCII) to potentially reduce constant factors. However, the HashSet is perfectly acceptable and more general for Unicode.

3. The while loop inside the for loop might be slightly inefficient in worst-case scenarios (e.g., when many duplicates are found), but overall it is optimal. Alternatively, you could use a HashMap to store the last occurrence index of each character to avoid the while loop, but that would require more space and might not be necessary.

4. Ensure that variable names are consistent: you use "left" and "right", which are clear, but in the comment you refer to "slow/left" and "fast/right" – this is fine, but make sure the code is self-explanatory.

Overall, great job!

VERDICT: PASS

PR #1194, opened 2026-03-28
working solution

avcode3:


super30admin:
### Custom Sort String (problem1.py)
Your solution is correct and efficient, running in linear time which is better than the reference solution that uses sorting (O(n log n)). Well done! 

Here are a few suggestions to improve code quality:
- Avoid using `map` as a variable name since it shadows the built-in function. Consider renaming it to `freq` or `char_count`.
- You can simplify the code by using `collections.Counter` for counting frequencies. This would make the code more concise and readable.
- Instead of using a list `final_str` and appending each character individually, you could build the string more efficiently by using `''.join()` with a list comprehension for the characters in `order`, and then appending the remaining characters. However, your current approach is acceptable.

Example using Counter:
```python

class Solution:
        freq = Counter(s)
        result = []
        for ch in order:
            if ch in freq:
                result.append(ch * freq[ch])
                del freq[ch]
        for ch, count in freq.items():
            result.append(ch * count)
        return ''.join(result)
```

This version is more concise and leverages the Counter class for clarity.

Overall, your solution is excellent. Keep up the good work!

VERDICT: PASS

---

###  Longest Substring without Repeat Characters (problem2.py)
Your solution is efficient and generally correct. It uses the sliding window technique with a hash map to achieve O(n) time complexity, which is optimal for this problem. Here are some points for improvement:

1. Variable Naming: Consider using more descriptive variable names. For example, 'slow_ptr' could be renamed to 'start' or 'left' to indicate the start of the current window. 'map_val' could be 'last_occurrence' to clarify that it stores the last index where each character was seen.

2. Edge Cases: Your solution handles most cases, but test it with an empty string and a string with all unique characters to ensure it works. For example, if s is "", the code returns 0, which is correct.

3. Explanation: Adding comments to explain the logic, especially why you use max(slow_ptr, map_val[ch]+1), would be helpful. This prevents the window start from moving backwards when encountering a character that was seen outside the current window.

4. Initialization: It's good practice to initialize variables explicitly. Here, 'max_val' is initialized to 0, which is correct.

5. Efficiency: Your solution is already efficient. No further optimizations are needed.

Overall, your solution is excellent and demonstrates a good understanding of the problem. Keep up the good work!

VERDICT: PASS

PR #1195, opened 2026-04-14
Cusotm sort string. Longest substring without repeating characters

paridhimalviya:
Custom Sort String (https://leetcode.com/problems/custom-sort-string/)
Longest Substring Without Repeating Characters(https://leetcode.com/problems/longest-substring-without-repeating-characters/)


super30admin:
### Custom Sort String (CustomSortString.swift)
Your solution is correct and efficient, with a linear time complexity that outperforms the sorting-based reference solution. Well done! Here are some suggestions for improvement:
1. Consider using more descriptive variable names for the dictionary, such as `charFrequency`, to make the code clearer.
2. Instead of appending each character individually in a loop, you could use `String(repeating:count:)` to append a character multiple times. For example, for a character `c` with frequency `f`, you can do `resultString += String(repeating: c, count: f)`. This would make the code more concise and potentially more efficient.
3. The second loop for the remaining characters can be simplified by iterating over the dictionary without explicitly removing keys. Since the dictionary is small, it's acceptable, but the removal is unnecessary. Alternatively, you can iterate over all 26 letters and append if the frequency is non-zero, but that might be less efficient.
4. Note that in Swift, strings are value types and each append might cause a copy. For better performance, you can use an array of characters to build the result and then convert to a string. However, for the constraints (s.length <= 200), it is acceptable.

Overall, your solution is excellent and demonstrates a good understanding of the problem.

VERDICT: PASS

---

###  Longest Substring without Repeat Characters (LongestSustringWithoutRepeatingCharacters.swift)
Note: The verdict should be based on the overall quality and correctness of the solution. If the solution is correct and meets the problem requirements, it should be a PASS. If there are issues with correctness, efficiency, or code quality, it should be NEEDS_IMPROVEMENT.

Let's begin.

VERDICT: NEEDS_IMPROVEMENT

PR #1196, opened 2026-04-23
Complete Strings-1

PrakarshKamal:


super30admin:
### Custom Sort String (CustomSortString.java)
Your solution is efficient and correct, which is great! You've achieved a linear time complexity, which is better than the reference solution that uses sorting. Here are a few suggestions for improvement:

1. **Variable Naming**: Consider using more descriptive variable names. For example, `ans` could be renamed to `result` or `output` for clarity.

2. **Efficiency in Loops**: In the final loop, you are iterating over `map.keySet()` and then calling `map.get(c)` for each key. This is acceptable but slightly inefficient because each `get` call is a constant-time operation, but it's still an extra lookup. You can iterate over the entry set instead to avoid this:
   ```java
   for (Map.Entry<Character, Integer> entry : map.entrySet()) {
        char c = entry.getKey();
        int freq = entry.getValue();
        for (int i = 0; i < freq; i++) {
            result.append(c);
        }
   }
   ```

3. **Code Comments**: Adding comments to explain each step of your algorithm would make the code more understandable for others.

4. **Consistency**: Ensure consistent use of data structures. While your use of HashMap is appropriate, note that since the problem only involves 26 characters, an array of size 26 could also be used for frequency counting, which might be slightly more efficient in terms of memory and speed. However, your current approach is acceptable.

5. **Edge Cases**: Your solution handles all edge cases correctly, including when `s` contains characters not in `order` and vice versa.

Overall, your solution is well-implemented and efficient. Keep up the good work!

VERDICT: PASS

---

###  Longest Substring without Repeat Characters (LongestSubstringWithoutRepeatingChars.java)
1. **Logical Error**: The while loop condition `while (set.contains(c))` is incorrect because it checks for the current right character 'c' before it is added to the set. Initially, 'c' is not in the set, so the while loop doesn't run when it should. Instead, you should check for duplicates after adding 'c', or restructure the code to add 'c' only if it's not present. However, the common approach is to check for duplicates before adding. Consider changing the condition to check for the presence of 'c' before adding, and then handle the shrinking accordingly.

2. **Correct Approach**: The typical sliding window solution uses a set and expands the right pointer. When a duplicate is found at 'right', you shrink the left pointer until the duplicate is removed. So, you should check if the set already contains 'c' (the current character) before adding it. If it does, then you remove the leftmost character and increment 'left' until the duplicate is removed. Then add 'c'. Your current code adds 'c' first and then checks for duplicates, which is backward.

3. **Efficiency**: Your current code has a worst-case time complexity of O(n²) due to the while loop potentially running for each right pointer. However, this is acceptable for the sliding window approach because each character is added and removed at most once, leading to an amortized O(n) time. But due to the logical error, it doesn't work correctly.

4. **Code Quality**: The code is well-structured and readable. Variable names are clear, and comments are helpful. However, the logic needs correction. Also, consider using a more efficient data structure like a HashSet is fine, but you can also use a boolean array of size 128 for ASCII characters for faster lookups and constant space.

5. **Optimization**: Instead of using a set, you can use a boolean array for ASCII characters (size 128) to mark presence. This would reduce constant factors and make the solution faster. Also, ensure that the while loop condition is fixed to correctly handle duplicates.

Suggested corrected code:
```java
    Set<Character> set = new HashSet<>();
    int maxLen = 0;
    int left = 0;
        char c = s.charAt(right);
        while (set.contains(c)) {
            set.remove(s.charAt(left));
            left++;
        }
        set.add(c);
        maxLen = Math.max(maxLen, right - left + 1);
    }
    return maxLen;
}
```

This code correctly checks for duplicates before adding the current character. The while loop ensures that if 'c' is already in the set, we shrink the window from the left until the duplicate is removed.

VERDICT: NEEDS_IMPROVEMENT

PR #1201, opened 2026-05-30
Create Problem1.java

YogeshPardeshi:
Problem1 - with Map solution to avoid one less pass

### DIFF
--- a/Problem1.java
+++ b/Problem1.java
@@ -1,0 +1,19 @@
+class Problem1 {
+    public int lengthOfLongestSubstring(String s) {
+        int slow = 0;
+        int n = s.length();
+        int max = 0;
+        HashMap<Character, Integer> map = new HashMap<>();
+
+        for (int i = 0; i < n; i++) {
+            char c = s.charAt(i);
+            if (map.containsKey(c)) {
+                slow = Math.max(slow, map.get(c) + 1);
+            }
+            map.put(c, i);
+            max = Math.max(max, i - slow + 1);
+        }
+
+        return max;
+    }
+}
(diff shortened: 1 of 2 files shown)


super30admin:
### Custom Sort String (Problem1.java)
**Strengths:**
- Correct algorithm that properly handles the custom sorting requirement
- Better time complexity (O(n + m)) compared to the reference solution's O(n log n)
- Clear and readable code structure
- Good use of StringBuilder for efficient string concatenation

**Areas for Improvement:**
- The `containsKey` check is redundant since `get` returns null for missing keys
- Could simplify the logic by using `getOrDefault` instead of separate `containsKey` and `get` calls
- Consider adding comments to explain the algorithm for better maintainability

**Minor optimization:**
```java
// Instead of:
if(map.containsKey(ch)) {
    int count = map.get(ch);
    ...
}

// Use:
int count = map.getOrDefault(ch, 0);
if (count > 0) {
    ...
}
```

VERDICT: PASS

---

###  Longest Substring without Repeat Characters (Problem2.java)
**Strengths:**
- Excellent use of the sliding window technique with HashMap - this is an optimal solution
- Clean and readable code structure
- Correctly handles edge cases like empty strings and all duplicate characters
- The `Math.max(i, map.get(ch)+1)` handles cases where the duplicate character's previous index is before the current window start

**Areas for Improvement:**
- Variable naming: `max` could be renamed to `maxLen` for better clarity
- Consider adding comments to explain the sliding window logic, especially the `Math.max(i, ...)` part which prevents shrinking the window when encountering a character seen earlier but outside current window
- The solution is missing the `import java.util.HashMap;` and `import java.util.Map;` statements (though these might be assumed in some contexts)

**Note:** The reference solution uses O(n²) brute force which is less efficient. The student's solution is actually superior in terms of time complexity.

VERDICT: PASS

